PR #5, opened 2022-02-04
Use -M0 instead of -frecord-sources

zanderso:
See discussion on https://github.com/flutter/engine/pull/31227

@chinmaygarde hopefully this is all you had in mind for this?

### DIFF
--- a/tools/build_metal_library.py
+++ b/tools/build_metal_library.py
@@ -88,7 +88,9 @@ def Main():
     command += [
       # Embeds both sources and driver options in the output. This aids in
       # debugging but should be removed from release builds.
-      "-frecord-sources",
+      # TODO(chinmaygarde): Use -frecord-sources when CI upgrades to
+      # Xcode 13.
+      "-MO",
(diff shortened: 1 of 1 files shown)


tools/build_metal_library.py
```diff
@@ -88,7 +88,9 @@ def Main():
     command += [
       # Embeds both sources and driver options in the output. This aids in
       # debugging but should be removed from release builds.
-      "-frecord-sources",
```
dnfield, line 91: Perhaps guard this with an xcode or clang version check?

dnfield: It seems like from Chinmay's comment it's more helpful

chinmaygarde (approved):

tools/build_metal_library.py
zanderso: I'd like for local builds to match CI. CI will be upgrading to Xcode 13 within weeks, so we can move both local and CI builds to `-frecord-sources` at the same time.

PR #6, opened 2022-02-04
Compensate for change to saveLayer

zanderso:
Trying to compensate for the change to `saveLayer()` in https://github.com/flutter/engine/pull/30957.

### DIFF
--- a/display_list/display_list_dispatcher.cc
+++ b/display_list/display_list_dispatcher.cc
@@ -137,8 +137,10 @@ static std::optional<Rect> ToRect(const SkRect* rect) {
 
 // |flutter::Dispatcher|
 void DisplayListDispatcher::saveLayer(const SkRect* bounds,
-                                      bool restore_with_paint) {
-  canvas_.SaveLayer(restore_with_paint ? paint_ : Paint{}, ToRect(bounds));
+                                      const flutter::SaveLayerOptions options) {
+  canvas_.SaveLayer(
+    options.renders_with_attributes() ? paint_ : Paint{},
+    ToRect(bounds));
 }
 
 // |flutter::Dispatcher|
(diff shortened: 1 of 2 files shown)


display_list/display_list_dispatcher.cc
```diff
@@ -137,8 +137,10 @@ static std::optional<Rect> ToRect(const SkRect* rect) {
 
 // |flutter::Dispatcher|
 void DisplayListDispatcher::saveLayer(const SkRect* bounds,
-                                      bool restore_with_paint) {
-  canvas_.SaveLayer(restore_with_paint ? paint_ : Paint{}, ToRect(bounds));
+                                      const flutter::SaveLayerOptions options) {
+  canvas_.SaveLayer(
+    options == flutter::SaveLayerOptions::kWithAttributes ? paint_ : Paint{},
```
flar, line 142: No, you don't want == here because there may be multiple flags set (do I even implement == on the Options?)

You want options.blah_uses_attributes_blah() which is a method that checks the flag and returns a boolean.

flar: If the "can do opacity optimization" flag was also set on this saveLayer then your test would fail.

flar (changes requested):

display_list/display_list_dispatcher.cc
flar: You want `options.renders_with_attributes()`

zanderso: Yeah, you have `==` on `SaveLayerOptions`: https://github.com/flutter/engine/blob/main/display_list/display_list.h#L182

So this should be `options.renders_with_attributes() ? paint_ : Paint{}`?

zanderso: Thanks! Changed to `options.renders_with_attributes()`.

flar (approved):
LGTM

PR #9, opened 2022-02-16
Add viewport and scissor support

bdero:
Used optionals for these because the default behavior with their absence will be similar across graphics backends.
Most modern APIs support multiple viewports and multiple scissor rects (not sure about driver coverage though), but GLES3 does not.

I think it's reasonable to include these as command params against all rendering backends. In Vulkan, we'd flip these on as [dynamic state](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkDynamicState.html) in pipelines and then set them with commands.

Used these in an [imgui backend](https://github.com/bdero/impeller/blob/3ac316755d016f7ea25c8e4504d5b5a151dd226a/bdero_pbr/imgui/imgui_impl_impeller.cc#L184-L188) that targets impeller 🙂 :

https://user-images.githubusercontent.com/919017/154174965-91656bd4-3bdf-408c-8643-94aaf117f362.mov

### DIFF
--- a/geometry/geometry_unittests.cc
+++ b/geometry/geometry_unittests.cc
@@ -335,5 +335,37 @@ TEST(GeometryTest, RectContainsPoint) {
   }
 }
 
+TEST(GeometryTest, RectContainsRect) {
+  {
+    Rect a(100, 100, 100, 100);
+    ASSERT_TRUE(a.Contains(a));
+  }
+  {
+    Rect a(100, 100, 100, 100);
+    Rect b(0, 0, 0, 0);
+    ASSERT_FALSE(a.Contains(b));
+  }
+  {
+    Rect a(100, 100, 100, 100);
+    Rect b(150, 150, 20, 20);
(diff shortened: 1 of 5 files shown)


chinmaygarde (approved):
So cool. I really like the concept of having a cheater GUI for adjusting rendering related options to aid in tuning. Using ImGUI for such a UI seems pretty cool even if it is during debugging. I do this in [my other side projects here](https://github.com/chinmaygarde/radar/blob/ee79b5bd39b356b35e5205a861c97c7c7d132f71/Library/Compositor/Source/Primitive/ColoredPathStrokePrimitive.cc#L38).

bdero:
I like what you have going on in radar with the inline statics -- very nice!
I suspected there would come a time where we'd want to optionally add something like this to the playground renderer for demos/easier debugging tests. With that in mind I'll rework it into a mergable state soon. :)

PR #32, opened 2022-02-25
Fix MoveTo for fills

dnfield:
Fixes flutter/flutter#99032

Related to https://github.com/flutter/impeller/pull/31

Tessellates individual contours of a polyline separately, to avoid treating `MoveTo` commands as introducing new contours. Removes the `DISABLED` from the test cases, since they now render the same as Skia.

### DIFF
--- a/renderer/tessellator.cc
+++ b/renderer/tessellator.cc
@@ -61,12 +61,24 @@ bool Tessellator::Tessellate(const Path::Polyline& contours,
   /// Feed contour information to the tessellator.
   ///
   static_assert(sizeof(Point) == 2 * sizeof(float));
-  ::tessAddContour(tessellator.get(),       // the C tessellator
-                   kVertexSize,             //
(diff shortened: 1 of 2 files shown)


renderer/tessellator.cc
```diff
@@ -61,12 +61,24 @@ bool Tessellator::Tessellate(const Path::Polyline& contours,
...
-  );
+  size_t start_point_index = 0;
+  for (size_t end_point_index : contours.breaks) {
+    ::tessAddContour(tessellator.get(),  // the C tessellator
+                     kVertexSize,        //
+                     contours.points.data() + start_point_index,  //
+                     sizeof(Point),                               //
+                     end_point_index - start_point_index          //
+    );
+    start_point_index = end_point_index;
```
dnfield, line 72: Hm. I guess this has some risk of buffer overflowing. I'll fix that.

bdero (approved):

PR #34, opened 2022-02-25
Remove duplicate points between connected components when generating polylines

bdero:
This PR is chained on top of https://github.com/flutter/impeller/pull/33.
Used in https://github.com/flutter/impeller/pull/35.

Removing duplicate points between connected components makes generating clean stroke geometry easier.

### DIFF
--- a/geometry/path.cc
+++ b/geometry/path.cc
@@ -197,9 +197,15 @@ bool Path::UpdateMoveComponentAtIndex(size_t index,
 Path::Polyline Path::CreatePolyline(
     const SmoothingApproximation& approximation) const {
   Polyline polyline;
-  auto collect_points = [&polyline](const std::vector<Point>& collection) {
-    polyline.points.reserve(polyline.points.size() + collection.size());
-    polyline.points.insert(polyline.points.end(), collection.begin(),
+  bool new_contour = true;
+  auto collect_points = [&polyline,
+                         &new_contour](const std::vector<Point>& collection) {
+    size_t offset = new_contour ? 0 : 1;
+    new_contour = false;
+
+    polyline.points.reserve(polyline.points.size() + collection.size() -
+                            offset);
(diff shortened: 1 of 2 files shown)


geometry/path.cc
```diff
@@ -197,9 +197,15 @@ bool Path::UpdateMoveComponentAtIndex(size_t index,
 Path::Polyline Path::CreatePolyline(
     const SmoothingApproximation& approximation) const {
   Polyline polyline;
-  auto collect_points = [&polyline](const std::vector<Point>& collection) {
-    polyline.points.reserve(polyline.points.size() + collection.size());
-    polyline.points.insert(polyline.points.end(), collection.begin(),
+  bool new_contour = true;
```
dnfield, line 200: Instead of doing this, what if we just made all the `CreatePolyline` calls only return the second point, and made `moveTo` calls use the destination point? I think that reads as slightly more straightforward, but maybe I'm missing something.

bdero: Including the first point of the next contour also needs to happen if the contour doesn't start with a MoveTo.

dnfield: I think Skia does move to insertion for that but this is probably fine too

bdero: That can only happen if it's the first contour on the path... lemme see if we're behaving the same as Skia in that case. I think it should start at 0, 0.

bdero: I like the idea of inserting a move at the beginning and started implementing it, but would prefer to land in a followup if you don't mind. I gotta make it overwrite back to back moves and stuff.

dnfield: SGTM

dnfield (approved):
LGTM. Consider adding a TODO for any code you plan to change in follow up with an issue filed for it.

bdero:
Thanks, appended a todo: https://github.com/flutter/flutter/issues/99177

PR #53, opened 2022-03-03
Move Contents classes into separate translation units

bdero:
I put everything under a `contents` directory, but not opposed to putting them somewhere else (or maybe inverting the header names for sorting purposes).

I tried to reduce the headers and forward declare when possible.
I think the biggest header explosion is `SolidStrokeContents` right now, since it includes `solid_stroke.vert.h` for the cap/join proc definitions. We can probably rework it to avoid this at some point.


chinmaygarde (approved):
Thank you!

PR #59, opened 2022-03-08
Remove FML dependency on geometry, tessellator

dnfield:
//flutter/fml depends on the Dart VM for _at least_ tracing and the implementation of `TimePoint::Now`. This is a very heavy dependency for the tessellation library, which really doesn't need it.

I've done the following:

- Dropped the FML dependency from `impeller_component`.
- Added the dependency to targets that still use it.
- Removed the dependency from `//flutter/impeller/geometry`, which only used it for some macros and DCHECKs
- Removed the dependency from `//flutter/impeller/tessellator`, which only used it for a single trace event.

### DIFF
--- a/geometry/path.cc
+++ b/geometry/path.cc
@@ -6,7 +6,6 @@
 
 #include <optional>
 
-#include "flutter/fml/logging.h"
 #include "impeller/geometry/path_component.h"
 
 namespace impeller {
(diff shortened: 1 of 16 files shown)


geometry/path.cc
```diff
@@ -19,8 +18,6 @@ Path::~Path() = default;
 
 std::tuple<size_t, size_t> Path::Polyline::GetContourPointBounds(
     size_t contour_index) const {
-  FML_DCHECK(contour_index < contours.size());
```
dnfield, line 22: Filed https://github.com/flutter/flutter/issues/99727. This method isn't safe as written anyway.

chinmaygarde (approved):

PR #65, opened 2022-03-08
[tessellator]Expose and document smoothing approximation via the C/Dart API. Minor refactor of tessellator to supply fill type to method rather than to object.

dnfield:
See https://github.com/flutter/flutter/issues/99769

This allows clients of the tessellation API to specify different smoothing values to obtain more vertices for curves that will be drawn at scale.

Adds some docs, updates a link, and moves the Tessellator fill type to the method (partly in preparation for a second method on Tessellator to tessellate strokes).

Updates the IMPELLER_API macro to work on Windows (although it's currently very difficult to build on windows since a bunch of the build rules lead to errors when not on macOS).

@jonahwilliams FYI


geometry/path_component.h
```diff
@@ -12,10 +12,31 @@
...
+  /// The scaling coefficient to use when translating to screen coordinates.
+  ///
+  /// Values approaching 0.0 will generate smoother looking curves with a
+  /// greater number of vertices, and will be more expensive to calculate.
   Scalar scale;
+
+  /// The tolerance value in radians for calculating sharp angles.
+  ///
+  /// Values approaching 0.0 will provide more accurate approximation of sharp
+  /// turns. A 0.0 vlaue means angle conditions are not considered at all.
```
jonahwilliams, line 30:
```suggestion
  /// turns. A 0.0 value means angle conditions are not considered at all.
```

tessellator/dart/lib/tessellator.dart
```diff
@@ -7,6 +7,53 @@ import 'dart:ffi' as ffi;
...
+  /// The scaling coefficient to use when translating to screen coordinates.
+  ///
+  /// Values approaching 0.0 will generate smoother looking curves with a
+  /// greater number of vertices, and will be more expensive to calculate.
+  final double scale;
+
+  /// The tolerance value in radians for calculating sharp angles.
+  ///
+  /// Values approaching 0.0 will provide more accurate approximation of sharp
+  /// turns. A 0.0 vlaue means angle conditions are not considered at all.
```
jonahwilliams, line 47:
```suggestion
  /// turns. A 0.0 value means angle conditions are not considered at all.
```

jonahwilliams:
I don't see the IMPELLER_API changes, did you mean to upload them in this PR?

geometry/path_component.h
dnfield: Ah, thanks, fixed these.

dnfield:
My push attempt got messed up and Ididn't notice. It should be there now.

chinmaygarde (approved):

PR #68, opened 2022-03-10
Prevent hanging when running back-to-back playground tests

bdero:
This has been bothering me for a while, so I decided to look into it today.

https://user-images.githubusercontent.com/919017/157591706-d73d30ae-babd-4676-a8d4-22f243aa65fa.mov

### DIFF
--- a/playground/playground.mm
+++ b/playground/playground.mm
@@ -107,10 +107,29 @@ bool Playground::OpenPlaygroundHere(Renderer::RenderCallback render_callback) {
   ImGui::StyleColorsDark();
   ImGui::GetIO().IniFilename = nullptr;
 
-  if (::glfwInit() != GLFW_TRUE) {
-    return false;
+  // This guard is a hack to work around a problem where glfwCreateWindow
+  // hangs when opening a second window after GLFW has been reinitialized (for
+  // example, when flipping through multiple playground tests).
+  //
+  // Explanation:
+  //  * glfwCreateWindow calls [NSApp run], which begins running the event loop
+  //    on the current thread.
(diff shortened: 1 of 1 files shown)


dnfield (approved):

playground/playground.mm
```diff
@@ -107,10 +107,29 @@ static void PlaygroundKeyCallback(GLFWwindow* window,
...
+  //  * GLFW then immediately stops the loop when applicationDidFinishLaunching
+  //    is fired.
+  //  * applicationDidFinishLaunching is only ever fired once during the
+  //    application's lifetime, so subsequent calls to [NSApp run] will always
+  //    hang with this setup.
+  //  * glfwInit resets the flag that guards against [NSApp run] being
+  //    called a second time, which causes the subsequent `glfwCreateWindow` to
+  //    hang indefinitely in the event loop, because
+  //    applicationDidFinishLaunching is never fired.
+  static bool first_run = true;
```
zanderso, line 126: Is there a more modern C++-y way to do this? Is this a good thing to use https://en.cppreference.com/w/cpp/thread/call_once ?

bdero: Maybe this?
```cpp
    static std::once_flag init_flag;
    std::call_once(init_flag, []() { ::glfwInit(); });
    if (::glfwGetError(nullptr) != GLFW_NO_ERROR) {
      return false;
    }
```
I don't think there's a primitive that can hide the state in some way, but this is one for the threading toolbox I'd say.

zanderso (approved):

chinmaygarde (approved):

PR #86, opened 2022-03-16
Don't use `Add[Thing]` operations when emulating Skia path components

bdero:
Don't use `Add[Thing]` operations when emulating Skia path components because they intentionally insert `MoveTo`s.

We should probably refactor this to skip around PathBuilder altogether and convert Skia's path components more or less directly instead of using the `PathBuilder`, which exposes a more nuanced/complex interface to emulate Skia's user path building functionality.


bdero:
image

dnfield (approved):
LGTM

bdero:
Also fixed the line component interpretation and I think we're good!

image

PR #87, opened 2022-03-17
Transform clips by the entity's transform

bdero:
Some print debugging and GPU snapshots later, it turns out that all clip paths are supposed to be done in the current transform space, not screen space.

This can also be seen in the docs: https://api.skia.org/classSkCanvas.html#abc79bd478516f398c31ae9f501154184

https://user-images.githubusercontent.com/919017/158735467-76ea3fe3-24b2-46e6-9ecf-7f650eee858a.mov

### DIFF
--- a/aiks/aiks_unittests.cc
+++ b/aiks/aiks_unittests.cc
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+#include <array>
 #include "flutter/testing/testing.h"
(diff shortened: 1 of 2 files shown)


dnfield (commented):
Change makes sense, can you add a test?

bdero:
Sure, added `AiksTest.ClipsUseCurrentTransform`:

image

dnfield (approved):
LGTM.

We're going to need some way to add more unit-testy type tests for things like this, pretty soon these playground tests will get unmanageable.

PR #91, opened 2022-03-18
Apply Aiks transforms in the canvas space, not world space

bdero:
This fixes the transform for everything except the background in the [page flip example](https://github.com/bizz84/page_flip_builder). I don't see any more transform-related issues in either page flip or gallery.

### DIFF
--- a/aiks/canvas.cc
+++ b/aiks/canvas.cc
@@ -61,7 +61,7 @@ bool Canvas::Restore() {
 }
 
 void Canvas::Concat(const Matrix& xformation) {
-  xformation_stack_.back().xformation = xformation * GetCurrentTransformation();
+  xformation_stack_.back().xformation = GetCurrentTransformation() * xformation;
 }
 
 void Canvas::ResetTransform() {
--- a/aiks/aiks_unittests.cc
+++ b/aiks/aiks_unittests.cc
@@ -132,9 +132,7 @@ TEST_F(AiksTest, ClipsUseCurrentTransform) {
 
   canvas.Translate(Vector3(300, 300));
   for (int i = 0; i < 15; i++) {
-    canvas.Translate(-Vector3(300, 300));
     canvas.Scale(Vector3(0.8, 0.8));
-    canvas.Translate(Vector3(300, 300));
 
     paint.color = colors[i % colors.size()];
     canvas.ClipPath(PathBuilder{}.AddCircle({0, 0}, 300).TakePath());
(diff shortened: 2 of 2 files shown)


aiks/canvas.cc
```diff
@@ -61,7 +61,7 @@ bool Canvas::Restore() {
 }
 
 void Canvas::Concat(const Matrix& xformation) {
-  xformation_stack_.back().xformation = xformation * GetCurrentTransformation();
+  xformation_stack_.back().xformation = GetCurrentTransformation() * xformation;
```
zanderso, line 64: Should matrix multiply use a different operator?

bdero: I think it's consistent with glsl/glm behavior. The mult operator looks like the right order for column major, but need to double check this.

bdero:
Ugh, I just spent hours debugging a transform problem I inadvertently introduced via my own instrumentation. This change alone actually _does_ solve all the transform problems in the page flip example. Checking what's going on with gallery now.

bdero:
Yup I think all the positioning looks correct in gallery with this.

aiks/canvas.cc
bdero: Yup, this looks consistent with the ordering convention. Canvas transform ops are supposed to make adjustments _within_ the current transform space rather than world space (as was being assumed in some tests).

dnfield (approved):

aiks/canvas.cc
flar: Since the mathematical `*` operator is supposed to be commutative, using it for a matrix multiply which is non-commutative seems like a potential source of confusion. (Is that what you were getting at @zanderso?)

flar: In particular, why was this PR needed to switch the operands of a mathematical multiply operator?

Rhetorical answer: because it wasn't really a mathematical multiply despite the notation...

dnfield: Pre or post multiplying a matrix is not commutative. Removing the operator is fine but if we keep it matrix multiplication is just special anyway right?

zanderso: > Since the mathematical `*` operator is supposed to be commutative, using it for a matrix multiply which is non-commutative seems like a potential source of confusion. (Is that what you were getting at @zanderso?)

Yeah, that's what I was getting at. Something that makes you think about the order might help avoid issues like this. I don't have strong feelings about what the replacement would be though.

bdero: I have a slight preference for keeping operator `*` for order dependent matrix-matrix and matrix-vector multiplication, as I think this is widely done/expected? (UE4, Unity, Godot, glsl/hlsl, [Flutter](https://api.flutter.dev/flutter/vector_math/Matrix4/operator_multiply.html), etc.)

flar: Hmmm. The world has already descended this slope I guess. If it's common and as long as developers keep in mind the row-vs-column notion of the matrices, then I suppose it is viable.

Although I would be more confident if the answer above was "of course this is the right order for column major matrices" rather than "let me go look this up to be sure"... ;-)

PR #92, opened 2022-03-20
switch dispatcher methods to new DlImageFilter objects

flar:
Enables engine PR https://github.com/flutter/engine/pull/32140

### DIFF
--- a/display_list/display_list_dispatcher.h
+++ b/display_list/display_list_dispatcher.h
@@ -67,7 +67,7 @@ class DisplayListDispatcher final : public flutter::Dispatcher {
   void setMaskFilter(const flutter::DlMaskFilter* filter) override;
 
   // |flutter::Dispatcher|
-  void setImageFilter(sk_sp<SkImageFilter> filter) override;
+  void setImageFilter(const flutter::DlImageFilter* filter) override;
 
   // |flutter::Dispatcher|
   void save() override;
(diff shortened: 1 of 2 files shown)


display_list/display_list_dispatcher.h
```diff
@@ -67,7 +67,7 @@ class DisplayListDispatcher final : public flutter::Dispatcher {
   void setMaskFilter(const flutter::DlMaskFilter* filter) override;
 
   // |flutter::Dispatcher|
-  void setImageFilter(sk_sp<SkImageFilter> filter) override;
+  void setImageFilter(const flutter::DlImageFilter* filter) override;
```
dnfield, line 70: I feel like I'm getting deja vu and I'm seeing this elsewhere so don't consider this blocking, but why aren't we using shared_ptr here in place of sk_sp?

flar: I think it is addressed in the base class (DlAttributeBase) where all of the design goals are listed for all of these objects.

The goal is to have a flat DL buffer that stands on its own and contains only data except for a few small cases where inlining is prohibitive (like Images). If you have to dispatch a shared_ptr to the Dispatcher, then you'd have to create a new instance inside a shared_ptr and hand that off. Most callees don't really need to keep the data around so it is not necessary for them to maintain shared ownership, so we provide an easy facility to get a shared copy if you need to keep ownership, but we don't impose that sharing expense on every dispatch call.

dnfield (approved):

chinmaygarde (approved):

PR #101, opened 2022-03-24
Render glyphs at the correct resolution with respect to the current transform

bdero:
I time boxed this and wrote a minimal solution because I was taking too long to reason about the best way to pipe this through the typographer stack.

This change scales up font metrics in accordance with the transform prior to glyph generation. Scaling by the max basis vector length of the transform ensures that glyphs are always be rendered at the right resolution to avoid pixelation against any affine space, even when heavily skewed.

### DIFF
--- a/aiks/canvas.cc
+++ b/aiks/canvas.cc
@@ -272,7 +272,7 @@ void Canvas::Save(bool create_subpass) {
 void Canvas::DrawTextFrame(TextFrame text_frame, Point position, Paint paint) {
   auto lazy_glyph_atlas = GetCurrentPass().GetLazyGlyphAtlas();
 
-  lazy_glyph_atlas->AddTextFrame(std::move(text_frame));
+  lazy_glyph_atlas->AddTextFrame(text_frame);
 
   auto text_contents = std::make_shared<TextContents>();
   text_contents->SetTextFrame(std::move(text_frame));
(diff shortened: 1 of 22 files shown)


bdero:
This dies in the gallery settings menu because a glyph lookup fails and I don't know why yet. I tried making the hashes/comparators less precise and that didn't fix it. Need to continue looking into it in a few hours.

bdero:
I tried a few more tricks today for this and failed to arrive at a perfect answer, so I removed the validation crash/added a debug log/filed a bug about how to reproduce the issue reliably: https://github.com/flutter/flutter/issues/100729
It's pretty hard to notice in gallery at regular speed, so I'm going to leave this to prioritize more noticeable visual issues right now.

bdero:
Alright, this is fixed now and there are no longer any failed lookups.

I took another swing at this and got the scale piped through in a more reasonable/lossless way. Now the font just has a scale field that the atlas renderer knows how to use (not very complicated). It's passed to the font at creation time since at that point we already know what the current transform will be, and `TextContents` doesn't need to behave any differently.

This is far less gross and is not really a hack anymore, so I think it's good to go!

bdero:
Screen Shot 2022-03-24 at 8 10 58 PM

chinmaygarde (approved):

PR #107, opened 2022-03-29
Add blur styles to gaussian blur

bdero:
Includes/relies on the changes in https://github.com/flutter/impeller/pull/103. ~Use [this diff view](https://github.com/bdero/impeller/compare/bdero/filter-input...bdero:bdero/blur-style) to view this PR without the dependency.~

https://user-images.githubusercontent.com/919017/160531260-a3f3081e-2c7d-41b9-9c57-e95ea647f71e.mov

This involved an arduous debug session, but some screenshots that are kinda aesthetically pleasing came out of it along the way!

Screen Shot 2022-03-28 at 2 46 01 PM

Screen Shot 2022-03-28 at 6 29 01 PM

### DIFF
--- a/entity/contents/filters/filter_contents.cc
+++ b/entity/contents/filters/filter_contents.cc
@@ -61,20 +61,26 @@ std::shared_ptr<FilterContents> FilterContents::MakeBlend(
 
 std::shared_ptr<FilterContents> FilterContents::MakeDirectionalGaussianBlur(
     FilterInput::Ref input,
-    Vector2 blur_vector) {
+    Vector2 blur_vector,
+    BlurStyle blur_style,
+    FilterInput::Ref source_override) {
   auto blur = std::make_shared<DirectionalGaussianBlurFilterContents>();
   blur->SetInputs({input});
   blur->SetBlurVector(blur_vector);
+  blur->SetBlurStyle(blur_style);
+  blur->SetSourceOverride(source_override);
   return blur;
 }
 
 std::shared_ptr<FilterContents> FilterContents::MakeGaussianBlur(
     FilterInput::Ref input,
     Scalar sigma_x,
-    Scalar sigma_y) {
-  auto x_blur = MakeDirectionalGaussianBlur(input, Point(sigma_x, 0));
-  auto y_blur =
-      MakeDirectionalGaussianBlur(FilterInput::Make(x_blur), Point(0, sigma_y));
+    Scalar sigma_y,
+    BlurStyle blur_style) {
+  auto x_blur =
+      MakeDirectionalGaussianBlur(input, Point(sigma_x, 0), BlurStyle::kNormal);
+  auto y_blur = MakeDirectionalGaussianBlur(
+      FilterInput::Make(x_blur), Point(0, sigma_y), blur_style, input);
   return y_blur;
 }
 
--- a/entity/contents/filters/gaussian_blur_filter_contents.cc
+++ b/entity/contents/filters/gaussian_blur_filter_contents.cc
@@ -30,6 +30,38 @@ void DirectionalGaussianBlurFilterContents::SetBlurVector(Vector2 blur_vector) {
   blur_vector_ = blur_vector;
 }
 
+void DirectionalGaussianBlurFilterContents::SetBlurStyle(BlurStyle blur_style) {
+  blur_style_ = blur_style;
+
+  switch (blur_style) {
+    case FilterContents::BlurStyle::kNormal:
+      src_color_factor_ = false;
+      inner_blur_factor_ = true;
+      outer_blur_factor_ = true;
+      break;
+    case FilterContents::BlurStyle::kSolid:
+      src_color_factor_ = true;
+      inner_blur_factor_ = false;
+      outer_blur_factor_ = true;
+      break;
+    case FilterContents::BlurStyle::kOuter:
+      src_color_factor_ = false;
+      inner_blur_factor_ = false;
+      outer_blur_factor_ = true;
+      break;
+    case FilterContents::BlurStyle::kInner:
+      src_color_factor_ = false;
+      inner_blur_factor_ = true;
+      outer_blur_factor_ = false;
+      break;
+  }
+}
+
+void DirectionalGaussianBlurFilterContents::SetSourceOverride(
+    FilterInput::Ref alpha_mask) {
+  source_override_ = alpha_mask;
+}
+
 bool DirectionalGaussianBlurFilterContents::RenderFilter(
     const FilterInput::Vector& inputs,
     const ContentContext& renderer,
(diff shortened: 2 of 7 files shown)


bdero:
Whoops, this is only really a thing for alpha mask blurs, which I'll most likely add as a separate filter.

bdero:
I've been thinking about the right way to do mask blurs across all contents types. After having a bit of a thonk about this, I realized these image filter blur styles have identical behavior to mask filter blurs over solid fills/strokes, which is very convenient and less tricky/expensive than alternate solutions that involved re-tessellating geometry or dealing with unpremultiplied source colors.

So for mask filter blurs over solid strokes/fills, we can use this path. For mask filter blurs over images (which is a different problem because the extended color is just the image border color), I will be adding a separate draw path.

entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -45,43 +77,59 @@ bool DirectionalGaussianBlurFilterContents::RenderFilter(
...
+  // LTRB
+  Scalar uv_src[4] = {
+      (filter_bounds.GetLeft() - source_bounds.GetLeft()) /
+          source_bounds.size.width,
+      (filter_bounds.GetTop() - source_bounds.GetTop()) /
+          source_bounds.size.height,
+      1 + (filter_bounds.GetRight() - source_bounds.GetRight()) /
+              source_bounds.size.width,
+      1 + (filter_bounds.GetBottom() - source_bounds.GetBottom()) /
+              source_bounds.size.height,
```
flar, line 112: I'm not clear on how these calculations work in a rotated coordinate system, but I don't think we'll encounter those in the near term anyway.

entity/shaders/gaussian_blur.frag
```diff
@@ -27,11 +32,9 @@ float Gaussian(float x) {
 }
 
 // Emulate SamplerAddressMode::ClampToBorder.
-vec4 SampleWithBorder(vec2 uv) {
-  if (uv.x > 0 && uv.y > 0 && uv.x < 1 && uv.y < 1) {
-    return texture(texture_sampler, uv);
-  }
-  return vec4(0);
+vec4 SampleWithBorder(sampler2D tex, vec2 uv) {
+  float within_bounds = float(uv.x > 0 && uv.y > 0 && uv.x < 1 && uv.y < 1);
```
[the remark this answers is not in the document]
flar: Ah, I see, it's emulating a specific sampling mode. Still, half open intervals are generally your friend...

entity/shaders/gaussian_blur.frag
```diff
@@ -40,8 +43,18 @@ void main() {
...
-  frag_color = total / total_gaussian;
+
+  vec4 blur_color = total / total_gaussian;
+
+  vec4 src_color = SampleWithBorder(alpha_mask_sampler, v_src_texture_coords);
+  float blur_factor = v_inner_blur_factor * src_color.a +
+                      v_outer_blur_factor * (1.0 - src_color.a);
+
+  frag_color =
+      blur_color * blur_factor + src_color * v_src_factor * src_color.a;
```
flar, line 59: First, doesn't src_color * src_color.a double premultiply the source color when we have src_factor?

Second, the inner part should always be the original, no additional alpha? The right half of this equation will (ignorig the potential double-premultiply) contribute the full interior color and the left half will contribute some more from the gaussian. That should over-apply the inner color for the case of an inner filter.

flar (commented):
The real question is how well do these equations match

https://fiddle.skia.org/c/87e545293e4322e0c7cf522ff54fdad3

flar:
I am also curious as to how these filters work with source images that have alpha in them. Does the inner blur leak into the interior in areas where the image evaluates as "alpha < 1"? In my Skia fiddle above I think the alpha I was using is applied post-mask-filter, so I don't think I truly demonstrated that case.

flar:
Here's another fiddle where I bake the alpha into the image rather than using a paint alpha to see how it affects the solid and inner blurs:

https://fiddle.skia.org/c/fd0866252906430ca9906d8d74737361

entity/shaders/gaussian_blur.frag
bdero: Good catch, removed. The added `src_color.a` factor was leftover from a previous (failed) attempt at making both passes use the same style option. We definitely don't want it.

> That should over-apply the inner color for the case of an inner filter.

`v_src_factor` is set to zero for `kInner`.

entity/contents/filters/gaussian_blur_filter_contents.cc
bdero: FWIW, this will be replaced when we add a [local transform](https://github.com/flutter/impeller/pull/103#discussion_r837987229) for filter inputs.

entity/contents/filters/filter_contents.cc
```diff
@@ -61,20 +61,26 @@ std::shared_ptr<FilterContents> FilterContents::MakeBlend(
...
     FilterInput::Ref input,
     Scalar sigma_x,
-    Scalar sigma_y) {
-  auto x_blur = MakeDirectionalGaussianBlur(input, Point(sigma_x, 0));
-  auto y_blur =
-      MakeDirectionalGaussianBlur(FilterInput::Make(x_blur), Point(0, sigma_y));
+    Scalar sigma_y,
+    BlurStyle blur_style) {
+  auto x_blur =
+      MakeDirectionalGaussianBlur(input, Point(sigma_x, 0), BlurStyle::kNormal);
```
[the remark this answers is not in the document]
bdero: For this linear two pass blur, the untampered blur output of the first pass needs to be fed into the second pass, and then the second pass applies the lossy alpha mask operation. If the first pass was kSolid, for example, any outer blur fragment in the second pass that samples from the non-blurred center of the first pass image would end up being the wrong color.

entity/shaders/gaussian_blur.frag
```diff
@@ -40,8 +43,18 @@ void main() {
   for (float i = -v_blur_radius; i <= v_blur_radius; i++) {
     float gaussian = Gaussian(i);
     total_gaussian += gaussian;
-    total += gaussian * SampleWithBorder(v_texture_coords +
-                                         v_blur_direction * i / v_texture_size);
+    total +=
+        gaussian * SampleWithBorder(texture_sampler,
+                                    v_texture_coords +
+                                        v_blur_direction * i / v_texture_size);
```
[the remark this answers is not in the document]
bdero: Done.

entity/shaders/gaussian_blur.frag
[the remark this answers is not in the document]
bdero: I think MSAA makes this a bit less noticeable, but yeah, I think the zero side should be inclusive here? Done.

bdero:
> I am also curious as to how these filters work with source images that have alpha in them.

This PR doesn't implement mask blurring for image inputs in the way Skia does -- it conveniently acts like a mask blur in the solid stroke/fill cases (because the color is uniform), but it doesn't for images. When mask blurring images, Skia extends the image with clamp border sampling and multiplies that factor into the alpha (see [this comment](https://github.com/flutter/impeller/pull/107#issuecomment-1085056945). I'll be adding this behavior in a follow-up.

I found Skia's behavior odd here at first, but it's kind of consistent: The mask blur filter extends the _geometry_ of the input and blurs the original _geometry mask_. It doesn't blur the alpha channel of the input image.

flar (approved):
My remaining concerns are in the subtle corner cases so I think this is a big first pass at this functionality. I'd be curious to check out test images to compare to the fiddles I posted.

Meanwhile, lgtm for now.

entity/shaders/gaussian_blur.frag
bdero: Ah, I see what you mean. When the mode is `kSource`, the inner blur should never mix with the source color. I replaced the src_color alpha factors with a strict binary mask instead to get this behavior for strokes/fills.

bdero: The blur factor is now this:
```cpp
  float blur_factor = v_inner_blur_factor * float(src_color.a > 0) +
                      v_outer_blur_factor * float(src_color.a == 0);
```

flar: That looks more like what I was expecting. Still approved...

PR #112, opened 2022-04-05
Fix solid stroke contour switching + round cap smoothing, and add transparent path overdraw example

bdero:
This test demonstrates a few solid stroke problems I'm investigating, but most notably: https://github.com/flutter/flutter/issues/101330

image

Corresponding Skia Fiddle:
https://fiddle.skia.org/c/027392122bec8ac2b5d5de00a4b9bbe2
image

### DIFF
--- a/aiks/aiks_unittests.cc
+++ b/aiks/aiks_unittests.cc
@@ -482,5 +482,43 @@ TEST_F(AiksTest, TransformMultipliesCorrectly) {
   // clang-format on
 }
 
+TEST_F(AiksTest, PathsShouldHaveUniformAlpha) {
+  // Compare with https://fiddle.skia.org/c/027392122bec8ac2b5d5de00a4b9bbe2
+  Canvas canvas;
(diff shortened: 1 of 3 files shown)


chinmaygarde (approved):

bdero:
Amended fixes for the two other solid stroke problems visible here:
* Append two transparent vertices at the start of the new contour instead of just one to prevent a thin triangle from sometimes being visible between contours.
* Make the smoothing approximation for round caps/joins incorporate the current transform.

Now it looks like this:
image

dnfield:
Erm.. this is still incorrect right? 

Skia guarantees that paths that self-intersect only hit pixels once. It seems like we're not doing that, right?

bdero:
Right, [this](https://github.com/flutter/flutter/issues/101330) is not fixed yet. The fixes I slid in were for the other minor issues I happened to notice while trying to demonstrate the overdraw.

PR #118, opened 2022-04-06
Standardize around "coverage" terminology; make coverage optional everywhere

bdero:
We already use the term "coverage" in `Entity` and `EntityPass`, which I hadn't noticed when I started working on filters and adding per-contents `GetBounds`. This change unifies the "area of the screen that may be affected" concept around the term "coverage".

Also, we distinguish between no coverage and empty coverage everywhere by returning optionals, which we were already doing in a few places (see also [this comment](https://github.com/flutter/impeller/pull/111#discussion_r842081668)).

### DIFF
--- a/entity/contents/contents.cc
+++ b/entity/contents/contents.cc
@@ -29,24 +29,27 @@ Contents::Contents() = default;
 
 Contents::~Contents() = default;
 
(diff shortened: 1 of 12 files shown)


chinmaygarde (approved):
I love it! Thank you for thinking of API UX and clarity.

PR #123, opened 2022-04-07
Standardize around blur sigma<->radius conversion factor

bdero:
Use the constant that Flutter + Skia use for the kernel radius/distribution sample cutoff: `1/sqrt(3)`. The cutoff is just barely visible. In practice, we're free to make this conversion factor whatever we want (with diminishing returns).
Making it 3, for example, would cause the kernel convolution to cover at least 99.9% of the gaussian distribution.

Also fixes a mistake in the gaussian function (use the correct constant `1/sqrt(2*pi)` instead of `1/(2*pi)`).

### DIFF
--- a/entity/contents/filters/gaussian_blur_filter_contents.cc
+++ b/entity/contents/filters/gaussian_blur_filter_contents.cc
@@ -24,6 +24,10 @@ DirectionalGaussianBlurFilterContents::
 
 void DirectionalGaussianBlurFilterContents::SetSigma(Sigma sigma) {
   if (sigma.sigma < kEhCloseEnough) {
+    // This cutoff is an implementation detail of the blur that's tied to the
+    // fragment shader. When the blur is set to 0, having a value slightly above
+    // zero makes the shader do 1 finite sample to pass the image through with
+    // no blur (while retaining correct alpha mask behavior).
     blur_sigma_ = Sigma{kEhCloseEnough};
     return;
   }
--- a/entity/contents/filters/gaussian_blur_filter_contents.cc
+++ b/entity/contents/filters/gaussian_blur_filter_contents.cc
@@ -22,12 +22,19 @@ DirectionalGaussianBlurFilterContents::DirectionalGaussianBlurFilterContents() =
 DirectionalGaussianBlurFilterContents::
     ~DirectionalGaussianBlurFilterContents() = default;
 
-void DirectionalGaussianBlurFilterContents::SetBlurVector(Vector2 blur_vector) {
-  if (blur_vector.GetLengthSquared() < kEhCloseEnough) {
-    blur_vector_ = Vector2(0, kEhCloseEnough);
+void DirectionalGaussianBlurFilterContents::SetSigma(Sigma sigma) {
+  if (sigma.sigma < kEhCloseEnough) {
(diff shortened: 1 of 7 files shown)


entity/shaders/gaussian_blur.frag
```diff
@@ -16,19 +16,19 @@ in vec2 v_texture_coords;
...
 
-const float kTwoPi = 6.283185307179586;
+const float kSqrtTwoPi = 2.50662827463;
 
 float Gaussian(float x) {
-  float stddev = v_blur_radius * 0.5;
-  float xnorm = x / stddev;
-  return exp(-0.5 * xnorm * xnorm) / (kTwoPi * stddev * stddev);
+  float variance = v_blur_sigma * v_blur_sigma;
+  return exp(-0.5 * x * x / variance) / (kSqrtTwoPi * v_blur_sigma);
```
flar, line 31: Confirmed that this is the standard gaussian blur equation as long as v_blur_sigma is non-negative.

entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -22,12 +22,19 @@ DirectionalGaussianBlurFilterContents::DirectionalGaussianBlurFilterContents() =
 DirectionalGaussianBlurFilterContents::
     ~DirectionalGaussianBlurFilterContents() = default;
 
-void DirectionalGaussianBlurFilterContents::SetBlurVector(Vector2 blur_vector) {
-  if (blur_vector.GetLengthSquared() < kEhCloseEnough) {
-    blur_vector_ = Vector2(0, kEhCloseEnough);
+void DirectionalGaussianBlurFilterContents::SetSigma(Sigma sigma) {
+  if (sigma.sigma < kEhCloseEnough) {
```
flar, line 26: Should this calculation be done in the Sigma constructor for safety? Or might there be different cutoffs for different situations?

flar (approved):
LGTM, left a comment on making it potentially more safe which I will leave up to you to decide.

bdero:
I added some comments that more rigorously describe `Sigma`, `Radius`, and our relationship between the two. I ended up inverting the constant and naming it like a unit conversion, as I find it's much more intuitive to explain/understand given the way we're using it and the consequences of changing it (`kBlurSigmaScale` is now `kKernelRadiusPerSigma`).

entity/contents/filters/gaussian_blur_filter_contents.cc
bdero: This cutoff is an implementation detail of the blur that's tied to the fragment shader. When the blur is set to 0, this makes the shader do 1 finite sample, which passes the image through with no blur (while retaining correct alpha mask behavior). We might do some different tricks in other filters that use gaussian distributions, so I'll leave this here for now.
I amended a comment detailing this behavior to make it clear that it more clear that it's implementation-specific.

bdero: Hmm, post-merge thoughts: This check should be done on the post-transformed blur vector, otherwise blurs with sigma=0 of objects zoomed > ~500x will end up slightly blurred with more than 1 sample. Will do that in a follow-up.

PR #126, opened 2022-04-08
Fix Path::GetBoundingBox crash for cubics

bdero:
Ran into this while making an override for the solid stroke coverage.

We incorporate the local min/max points for cubics by finding the roots of the first derivative of the x/y component cubics, but this doesn't end up enclosing the first/last point in most cases.

Demo of the issue (with the optional check to prevent the crash):

https://user-images.githubusercontent.com/919017/162394845-7b923333-0b32-47dc-867e-45a426367ebb.mov

The test I added creates a cubic with no derivative roots (it's just a line), which was causing `Path::GetBoundingBox` to crash before this change.


flar (approved):
LGTM

PR #127, opened 2022-04-08
Add solid stroke coverage override

bdero:
Relies on the `Path::GetBoundingBox` fix in https://github.com/flutter/impeller/pull/126.

Currently, the stroke coverage is path coverage, and so it always gets cut off when used with SaveLayer or Filters without this change.

To get comprehensive coverage, it's necessary to take the transformed stroke width, `Cap::kButt`, and the miter limit into account.

Demo of the updated playground test (green rectangles are the coverage):

https://user-images.githubusercontent.com/919017/162400554-4f099872-4286-4326-a231-8224aa19c81e.mov

### DIFF
--- a/entity/contents/solid_stroke_contents.cc
+++ b/entity/contents/solid_stroke_contents.cc
@@ -46,11 +46,9 @@ std::optional<Rect> SolidStrokeContents::GetCoverage(
   Vector2 max_radius_xy = entity.GetTransformation().TransformDirection(
       Vector2(max_radius, max_radius) * stroke_size_);
 
-  return Rect(
-      Point(std::floor(path_coverage->origin.x - max_radius_xy.x),
-            std::floor(path_coverage->origin.y - max_radius_xy.y)),
-      Size::Ceil(Size(path_coverage->size.width + max_radius_xy.x * 2,
-                      path_coverage->size.height + max_radius_xy.y * 2)));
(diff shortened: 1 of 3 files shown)


entity/contents/solid_stroke_contents.cc
```diff
@@ -27,6 +29,30 @@ const Color& SolidStrokeContents::GetColor() const {
...
+    max_radius = max_radius * kSqrt2;
+  }
+  if (join_ == Join::kMiter) {
+    max_radius = std::max(max_radius, miter_limit_ * 0.5f);
+  }
+  Vector2 max_radius_xy = entity.GetTransformation().TransformDirection(
+      Vector2(max_radius, max_radius) * stroke_size_);
+
+  return Rect(
+      Point(std::floor(path_coverage->origin.x - max_radius_xy.x),
```
flar, line 50: Why are you doing a floor/ceil here? Shouldn't it be up to the caller whether they want the bounds to be rounded out?

flar (approved):
LGTM with a thought on whether the coverage should be rounded out (though that may be per the contract of Entities. Still, if each entity does rounding out then we end up with compounded rounding if they are chained.

entity/contents/solid_stroke_contents.cc
bdero: Good point, thanks -- removed this.

PR #130, opened 2022-04-12
Filters: Replace Snapshot position with a full transform

bdero:
Replace the `Snapshot` position (which held the position offset of the texture) with a transformation matrix, enabling raw texture `FilterInput`s to be consumed by the graph without pre-passes. The transform is respected equally for any input variant (`Contents` or `Texture`). This also allows for optimizing some less common situations. For example, [MatrixTransform](https://api.skia.org/classSkImageFilters.html#a81f98126fe53bc0bec517d356e805057) filters could just return the same snapshot (sampling mode permitting) except with a multiplied input texture.

Most of the changes here are to make the existing filters respect the transform:
* For basic blends, each input is drawn using a separate draw call, and so the transform is just applied to the geometry.
* For advanced blends and the Gaussian blur, the geometry remains unaffected. Instead, input UVs are computed by transforming the filter's screen space coverage to each input's UV space.

### DIFF
--- a/entity/contents/contents.cc
+++ b/entity/contents/contents.cc
@@ -33,7 +33,7 @@ std::optional<Rect> Contents::GetCoverage(const Entity& entity) const {
   return entity.GetPathCoverage();
 }
 
-std::optional<Snapshot> Contents::RenderToTexture(
+std::optional<Snapshot> Contents::RenderToSnapshot(
     const ContentContext& renderer,
     const Entity& entity) const {
   auto bounds = GetCoverage(entity);
@@ -58,7 +58,8 @@ std::optional<Snapshot> Contents::RenderToTexture(
     return std::nullopt;
   }
 
-  return Snapshot{.texture = texture, .position = bounds->origin};
+  return Snapshot{.texture = texture,
+                  .transform = Matrix::MakeTranslation(bounds->origin)};
 }
 
 }  // namespace impeller
--- a/entity/contents/contents.h
+++ b/entity/contents/contents.h
@@ -39,11 +39,11 @@ class Contents {
   /// @brief Get the screen space bounding rectangle that this contents affects.
   virtual std::optional<Rect> GetCoverage(const Entity& entity) const;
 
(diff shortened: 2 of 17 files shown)


bdero:
I'm currently waiting for this PR to go in before writing more filters (to avoid building a house of cards, as changes to the approach in this PR may require significant changes in https://github.com/flutter/impeller/pull/132).

dnfield (approved):

bdero:
There's a problem with this design I encountered while working on the mask blur that I'd like to address in a followup hopefully before the end of the week: This transform is pretty important for computing coverage correctly on stuff like blurs, since the blur radius needs to be scaled by this matrix to know how much the coverage needs to be extended. Since this matrix is stored in the FilterInput snapshot, the snapshot needs to be resolved just for computing bounds. This doesn't really matter right now since we aren't doing any filter elision yet, but there's no reason this matrix needs to be in the snapshot -- it should really just be a local transform (relative to the entity transform) stored as a member in `FilterContents` itself.

Here's what I'm planning:
- Make the transform local (relative to the entity transform) instead of including the entity transform (we need to use it in both ways, so we should actually just store it in whatever way is most commonly used and add helpers to get both in `FilterContents`).
- Make `FilterContents` store the local transform as a member (rather than having it in the `FilterInput` snapshot) -- at this point, maybe just remove `GetSnapshot` from FilterInput and just have a `GetTexture`, `GetLocalTransform`, and `GetCoverage` directly accessible on the FilterInput that does the right thing.
- Add a special `FilterInput` variant for `FilterContents` -- maybe just subclass `FilterContents` at this point, since the code for each variant is pretty unique.
- Add methods to `FilterInput` for resolving the local or full screen space transforms of the input -- use these for computing coverage in filters.
- At some point: Add an optional pre-render virtual method to `FilterContents` that can be used to 1. manipulate the local transform, and 2. decide whether or not a render is necessary, enabling passthrough/filter elision/transform filter behaviors.

PR #132, opened 2022-04-13
Border mask blur

bdero:
Directly depends on changes in https://github.com/flutter/impeller/pull/130.

Adds a constant time mask blur for image borders.

This mask blur extends the geometry of the source image (with clamp border sampling) and applies a Gaussian blur to the original geometry mask.

The blur itself works by mapping the Gaussian distribution's indefinite integral (with a very close erf approximation) to the 4 edges of the UV rectangle and multiplying them.

https://user-images.githubusercontent.com/919017/163268514-cc9ea43a-0f73-4b00-b10a-dc490cc01a6f.mov

I'm planning a follow-up that allows filters to return snapshots, which will allow filters that don't modify coverage to do filtering into textures of the same size without having to worry about repositioning geometry/expanding contents. This way, we can chain `BorderMaskBlurFilterContents` after things like `BlendFilterContents` or the upcoming `ColorFilterContents` and it'll always extend the border of the original texture correctly. This would also enable filters to just pass through snapshots whenever no computation needs to be done.

### DIFF
--- a/entity/BUILD.gn
+++ b/entity/BUILD.gn
@@ -20,6 +20,8 @@ impeller_shaders("entity_shaders") {
     "shaders/texture_blend_screen.vert",
     "shaders/gaussian_blur.frag",
     "shaders/gaussian_blur.vert",
+    "shaders/border_mask_blur.frag",
+    "shaders/border_mask_blur.vert",
     "shaders/texture_fill.frag",
     "shaders/texture_fill.vert",
     "shaders/glyph_atlas.frag",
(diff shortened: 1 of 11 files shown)


bdero:
I'm currently waiting for this PR to go in before writing more filters (to avoid building a house of cards, as changes to the approach in this PR may require significant changes in https://github.com/flutter/impeller/pull/132).

chinmaygarde (approved):
Thanks. The fragment sl implementation was a particularly fun review!

PR #134, opened 2022-04-13
Move rendering TUs behind own flag

dnfield:
Fixes build issues seen in https://github.com/flutter/engine/pull/32658

Will require one change when rolling into the engine to guard header includes behind this new flag instead of the more generic one.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -20,38 +20,48 @@ config("impeller_public_config") {
 
 group("impeller") {
   public_deps = [
-    "aiks",
     "archivist",
     "base",
-    "display_list",
-    "entity",
     "geometry",
-    "image",
-    "renderer",
     "tessellator",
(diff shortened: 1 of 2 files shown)


BUILD.gn
```diff
@@ -20,38 +20,48 @@ config("impeller_public_config") {
...
-    "entity",
     "geometry",
-    "image",
-    "renderer",
     "tessellator",
-    "typographer",
   ]
+
+  if (impeller_supports_rendering) {
+    deps += [
```
bdero, line 30: Shouldn't this be `public_deps`?

bdero (approved):
LGTM if it builds.

BUILD.gn
dnfield: Yes

zanderso (approved):

PR #135, opened 2022-04-13
Build fixes for roll

dnfield:
For https://github.com/flutter/engine/pull/32658

### DIFF
--- a/renderer/backend/metal/command_buffer_mtl.mm
+++ b/renderer/backend/metal/command_buffer_mtl.mm
@@ -53,6 +53,18 @@ static CommandBuffer::Status ToCommitResult(MTLCommandBufferStatus status) {
   return CommandBufferMTL::Status::kError;
 }
 
+// TODO(dnfield): remove this declaration when we no longer need to build on
+// machines with lower SDK versions than 11.0.s
+#if !defined(MAC_OS_X_VERSION_11_0) || MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_X_VERSION_11_0
+typedef enum MTLCommandEncoderErrorState : NSInteger {
+  MTLCommandEncoderErrorStateUnknown = 0,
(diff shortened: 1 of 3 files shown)


dnfield:
Actually I don't think I need this.

dnfield:
The new define isn't essential, but the missing build dep and enum definition will make this landable.

bdero (approved):
If the new define/version check isn't needed, can we remove it? I don't understand it, but I will speculatively approve this PR. (I slipped the include fix in one of my PRs that still waiting on review BTW)

dnfield:
We can remove it when we're certain that no bot will build  using Xcode 12. Until then, the bots may be using an older macOS SDK that does not have the enum defined - so we conditionally define it to allow the TU to compile.

@jmagman FYI

dnfield:
Argh. I too eagerly merged this. the typedef isn't set up right

renderer/backend/metal/command_buffer_mtl.mm
```diff
@@ -53,6 +53,18 @@
   return CommandBufferMTL::Status::kError;
 }
 
+// TODO(dnfield): remove this declaration when we no longer need to build on
+// machines with lower SDK versions than 11.0.s
+#if !defined(MAC_OS_X_VERSION_11_0) || MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_X_VERSION_11_0
```
jmagman, line 58: What exactly are you trying to do?  Does `API_AVAILABLE(macos(11.0))` work?  Do you need an iOS availability constraint as `MTLCommandEncoderErrorStateToString` has `API_AVAILABLE(ios(14.0), macos(11.0))`

dnfield: See more of this continued in #136.

I'm trying to make sure that if we end up on a bot that has an older SDK that does not define this enum, we can still compile.

PR #136, opened 2022-04-13
Fix enum typedef for older mac SDKs

dnfield:
I .. I forget how to write ObjC.

I'm going to let CI finish on https://github.com/flutter/engine/pull/32658 (which currently points to my fork) before landing this.

### DIFF
--- a/renderer/backend/metal/command_buffer_mtl.mm
+++ b/renderer/backend/metal/command_buffer_mtl.mm
@@ -55,7 +55,7 @@ static CommandBuffer::Status ToCommitResult(MTLCommandBufferStatus status) {
 
 // TODO(dnfield): remove this declaration when we no longer need to build on
 // machines with lower SDK versions than 11.0.s
-#if !defined(MAC_OS_X_VERSION_11_0) || MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_X_VERSION_11_0
+#if !defined(MAC_OS_VERSION_11_0) || MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_VERSION_11_0
 typedef NS_ENUM(NSInteger, MTLCommandEncoderErrorState) {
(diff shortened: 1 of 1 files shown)


renderer/backend/metal/command_buffer_mtl.mm
```diff
@@ -55,8 +55,8 @@
 
 // TODO(dnfield): remove this declaration when we no longer need to build on
 // machines with lower SDK versions than 11.0.s
-#if !defined(MAC_OS_X_VERSION_11_0) || MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_X_VERSION_11_0
-typedef enum MTLCommandEncoderErrorState : NSInteger {
+#if !defined(MAC_OS_VERSION_11_0) || MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_VERSION_11_0
```
jmagman, line 58: All Mac engine builders except for the `Mac Web Engine` are running Xcode 13 (which has `MAC_OS_VERSION_11_0` defined)
https://github.com/flutter/engine/blob/289368cadccbce88913b3ab279ed754d705740f9/.ci.yaml#L73

`Mac Web Engine` Xcode migration blocked on https://github.com/flutter/flutter/issues/99885
https://github.com/flutter/engine/blob/289368cadccbce88913b3ab279ed754d705740f9/.ci.yaml#L327

dnfield: Need a guard for the macOS 12 defines too

dnfield:
With this patch applied all impeller related failures on flutter/engine#32658 are resolved.

renderer/backend/metal/command_buffer_mtl.mm
dnfield: So once Mac Web engine rolls we can remove the first block here - we won't be able to remove the second until we further upgrade.

zanderso (approved):

PR #137, opened 2022-04-14
Move static methods to anonymous namespace

dnfield:
Fixes symbol exposure that resulted in revert in https://github.com/flutter/engine/pull/32677

Apple Clang doesn't hide these even though Fuchsia Clang does.

The diff is confusing because I re-arranged some code, but this did not actually delete any lines.


dnfield:
I manually verified that when building this locally with Xcode and bitcode enabled, the symbols are exposed before this change and _not_ exposed after.

zanderso (approved):

PR #142, opened 2022-04-18
Implement mask blur in display list dispatcher

bdero:
We have 2 methods for achieving mask blur behavior:
- For images, use the BorderMaskBlurFilter, which is constant time per-fragment.
- For solid color fills, do a Gaussian blur, which is linear per-fragment, but does two subpasses. See below for future optimization ideas.

Future optimizations:
- Do `BorderMaskBlur` directly in the parent pass. Depending on the entity context, rendering into a subpass is unnecessary. This could also be done for the second gaussian blur pass. This is a feature I've been thinking about adding to `FilterContents`.
- For solid fills, trade GPU memory/time for CPU time: For the arbitrary path case, generate inset/outset geometry and give them distance coordinates. This isn't trivial for concave paths segments, but off the top of my head, I can think of a tunable heuristic to do this path-size linearly. The tunable parameter would tradeoff CPU time (computing more geometry) with overlapping penumbra artifact size.

kOuter mask blurred solid path over a kNormal mask blurred image:
Screen Shot 2022-04-18 at 2 24 03 PM

### DIFF
--- a/aiks/canvas.cc
+++ b/aiks/canvas.cc
@@ -237,6 +237,7 @@ void Canvas::DrawImageRect(std::shared_ptr<Image> image,
   contents->SetTexture(image->GetTexture());
   contents->SetSourceRect(source);
   contents->SetSamplerDescriptor(std::move(sampler));
+
   Entity entity;
   entity.SetPath(PathBuilder{}.AddRect(dest).TakePath());
   entity.SetBlendMode(paint.blend_mode);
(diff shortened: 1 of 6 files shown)


bdero:
Added text to the mix. When glyphs are rendered, they overlap. So to make it work right as a filter input, I changed the default `Contents::RenderToSnapshot` behavior to use `kSourceOver` instead of `kSource`.

image

dnfield (approved):

chinmaygarde (approved):

PR #150, opened 2022-04-21
Remove dependency on Dart, add crash handler to impellerc

dnfield:
Depends on https://github.com/flutter/engine/pull/32846

Fixes https://github.com/flutter/flutter/issues/101866

Added the crash handler because at one point changes in the engine I made caused impellerc to segfault and I had no idea why :)


chinmaygarde (approved):
Awesome. Thanks.

PR #153, opened 2022-04-22
implement drawDisplayList in impeller dispatcher

flar:
This is the first half of a much simpler fix for https://github.com/flutter/flutter/issues/100983


bdero (approved):
LGTM. I've never seen this get called before, but I assume the other half of the fix for [this](https://github.com/flutter/flutter/issues/100983) is to start using this for child display lists as opposed to just flattening them during execution, right?

flar:
> LGTM. I've never seen this get called before, but I assume the other half of the fix for [this](https://github.com/flutter/flutter/issues/100983) is to start using this for child display lists as opposed to just flattening them during execution, right?

Exactly. It should be a one line change to the engine, modulo dealing with inherited opacity which will be much simpler when the DlPaint PR lands.

PR #162, opened 2022-04-26
Set path for linear gradient contents; don't fail renders for failed tessellations

bdero:
Fixes a problem introduced in #145 which causes the frame not to render whenever there's a linear gradient drawn due to no path being set. This code is going to change a lot once we start landing the other color sources, so this is a minimal change to just get us out of the water.

Also, failed tessellations (which may happen due to user-supplied empty paths) shouldn't fail the whole frame, so I changed render methods to return `true` in this situation (which we already do in some places).

Follow up: It was kind of tedious to work out where the failed render was coming from. I'm thinking about going around and adding an error message for anything that will fail the frame/cancel rendering (that means anything that returns `false` in a render method) to make stuff like this easier to trace.

### DIFF
--- a/aiks/paint.cc
+++ b/aiks/paint.cc
@@ -11,6 +11,7 @@ namespace impeller {
 std::shared_ptr<Contents> Paint::CreateContentsForEntity(Path path,
                                                          bool cover) const {
   if (contents) {
+    contents->SetPath(std::move(path));
     return contents;
   }
 
--- a/entity/contents/linear_gradient_contents.cc
+++ b/entity/contents/linear_gradient_contents.cc
@@ -63,7 +63,7 @@ bool LinearGradientContents::Render(const ContentContext& renderer,
                                    vertices_builder.AppendVertex(vtx);
                                  });
 
-    if (result == Tessellator::Result::kFatalError) {
+    if (result == Tessellator::Result::kTessellationError) {
       return false;
     }
     return true;
--- a/entity/contents/linear_gradient_contents.cc
+++ b/entity/contents/linear_gradient_contents.cc
@@ -4,6 +4,7 @@
 
 #include "linear_gradient_contents.h"
 
+#include "flutter/fml/logging.h"
 #include "impeller/entity/contents/content_context.h"
 #include "impeller/entity/entity.h"
 #include "impeller/renderer/render_pass.h"
@@ -16,6 +17,7 @@ LinearGradientContents::LinearGradientContents() = default;
 LinearGradientContents::~LinearGradientContents() = default;
 
 void LinearGradientContents::SetPath(Path path) {
+  path_set_ = true;
   path_ = std::move(path);
 }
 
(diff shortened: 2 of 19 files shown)


aiks/paint.cc
```diff
@@ -11,6 +11,7 @@ namespace impeller {
 std::shared_ptr<Contents> Paint::CreateContentsForEntity(Path path,
                                                          bool cover) const {
   if (contents) {
+    contents->SetPath(std::move(path));
```
dnfield, line 14: Please add a DCHECK that asserts that `SetPath` has been called before `Render`.

entity/contents/linear_gradient_contents.cc
```diff
@@ -59,7 +59,7 @@ bool LinearGradientContents::Render(const ContentContext& renderer,
                                    vertices_builder.AppendVertex(vtx);
                                  });
     if (!result) {
-      return false;
+      return true;
```
dnfield, line 62: We need to distinguish here whether we failed because the path is empty or we failed because we ran out of memory or something. If we ran out of memory we really need to stop so that the process has as much of a chance as possible to allocate what it can to log a crash etc.

One way to do this would be to create a `TessellationResult` enum, and have it return early in the tessellator if the polyline is empty with an appropriate status.

dnfield (changes requested):

aiks/paint.cc
bdero: Done. Also been thinking about adding an `IsValid` for contents in general, but will save that for another day.

entity/contents/linear_gradient_contents.cc
bdero: Done. Also added a test unit for the tessellator.

entity/contents/linear_gradient_contents.cc
```diff
@@ -46,6 +48,8 @@ std::optional<Rect> LinearGradientContents::GetCoverage(
 bool LinearGradientContents::Render(const ContentContext& renderer,
                                     const Entity& entity,
                                     RenderPass& pass) const {
+  FML_DCHECK(path_set_) << "Render was called without setting a path.";
```
dnfield, line 51: Here and elsewhere:

Can we just check that the path is not empty (`FML_DCHECK(path_.GetComponentCount() > 0)`)? And then guard upstream so that even if Flutter app asks us to draw an empty path, we short circuit before we get to rendering contents?

That eliminates a weird bool that we're likely to forget to set somewhere, and that doesn't quite cover validity for contents.

bdero: Ehhh, yeah the extra state is ugly -- I reversed it. Thinking this through, I'm not so sure we really need any checks here. I think just having `EntityPass` print a message that says the render failed because of an entity failing would have made this issue easy enough to track down.

The original checks help to prevent a developer error, but having an empty path supplied here is not an error and shouldn't cause a failure here after the changes I made to the tessellator.

dnfield: I'm of a mind to say that we should make rendering an empty path an error at this point in the stack, but that could be hashed out in a separate PR.

dnfield (approved):
LGTM. We can figure out what to do with the API around "rendering" empty paths separately.

